Issue #1, opened 2020-09-01
Initial module setup


michaelgoin:
[NOTE]: # ( ^^ Provide a general summary of the request in the title above. ^^ )

## Summary

Setup module for use/development

* npm init
* package.json details
* project structure
* etc.

michaelgoin:
recommend starting with files allow list

PR #15, opened 2020-09-01
Initial module setup

michaelgoin:
Gets some initial scaffolding in place to start from.

Closes: https://github.com/newrelic/newrelic-node-apollo-server-plugin/issues/1

Here's what the current setup will look like packaged:

```
npm notice 📦  @newrelic/apollo-server-plugin@0.1.0
npm notice === Tarball Contents === 
npm notice 11.4kB LICENSE             
npm notice 199B   lib/create-plugin.js
npm notice 379B   index.js            
npm notice 1.0kB  package.json        
npm notice 2.5kB  README.md  
```


package.json
```diff
@@ -0,0 +1,32 @@
...
+    "url": "https://github.com/newrelic/newrelic-node-apollo-server-plugin.git"
+  },
+  "author": "New Relic Node.js agent team <nodejs@newrelic.com>",
+  "license": "Apache-2.0",
+  "bugs": {
+    "url": "https://github.com/newrelic/newrelic-node-apollo-server-plugin/issues"
+  },
+  "homepage": "https://github.com/newrelic/newrelic-node-apollo-server-plugin#readme",
+  "engines": {
+    "node": ">=10.0.0"
```
michaelgoin, line 22: Plan to start out 10+ to reduce future churn here.

.gitignore
```diff
@@ -0,0 +1,61 @@
...
+coverage
+
+# nyc test coverage
+.nyc_output
+
+# Grunt intermediate storage (http://gruntjs.com/creating-plugins#storing-task-files)
+.grunt
+
+# Bower dependency directory (https://bower.io/)
+bower_components
```
carlonewrelic, line 27: Are we planning on using Grunt and Bower?

michaelgoin: not particularly, no.

carlonewrelic: Should we just drop these lines then?

michaelgoin: 🤷 copy pasted. didn't think necessary to go through and scrub each line. if we want to get real picky, we can start from scratch and not grab from previous projects. when it comes to ignore files, i dont' care as much.

package.json
```diff
@@ -0,0 +1,36 @@
...
+  "main": "/src/index.js",
+  "scripts": {
+    "lint": "eslint *.js lib tests",
+    "test": "npm run unit",
+    "unit": "tap --test-regex='(\\/|^tests\\/unit\\/.*\\.test\\.js)$' --no-coverage"
+  },
+  "files": [
+    "index.js",
+    "lib/*.js"
+  ],
```
michaelgoin, line 14: @carlonewrelic added 'files' declaration which will limit files that get packaged. adding an example of the output to the main PR description. we shouldn't need an npmignore at any point on this repo... we can tweak this allow list, as needed.

carlonewrelic: Awesome!  💯

Issue #4, opened 2020-09-01
Setup CI in GitHub Actions


michaelgoin:
[NOTE]: # ( ^^ Provide a general summary of the request in the title above. ^^ )

## Summary

We may need one or more tests before this is actually possible. I believe tap yells at you if you give it a glob that resolves to nothing, unlike mocha.

PR #16, opened 2020-09-02
Adds initial GitHub Actions CI.

michaelgoin:
Closes: https://github.com/newrelic/newrelic-node-apollo-server-plugin/issues/4


astormnewrelic (approved):
Does the thing.

PR #21, opened 2020-09-10
Initial transaction naming

michaelgoin:
* Adds initial transaction naming 
* Adds versioned testing for apollo-server and apollo-server-express
* Adds versioned tests to CI

This PR does not yet add the base segment/span.

This isn't expected to necessarily be the final form of transaction names. I have some reasons for these pieces in this loose display but there will be further discussion needed and further understanding of impacts/standards of naming for the back-end and UI.

This will however give us a starting point including tests to cover several permutations of considerations. It will also give some initial structure to the plugin to start to reduce merge collisions going forward (still will be a chance) and testing structure as an example for future work.

I've purposely left some TODO's in the code that we need to resolve as we iterate. Some of which will require agent changes.

### Transaction Naming Rules

`[type]`: refers the operation type (query/mutation)
`[name]`: refers to the name, if provided, of the query. `mutation MyMutationName {...}`

#### Operation resolves: parsing and validation succeeded

Transaction is named off of the resolved operation plus the longest path derived from resolvers (willResolveField) hit.

`apollo-server/[type] [name | <anonymous>] [longest resolved path]

#### Validation fails: parsing succeeded but could not validate (invalid field for example)

Transaction is named off of the document definition parsed from the query/request. Essentially, what the user attempted. This is a bit more expensive but gives at least some strong clues to what went wrong. Once fixed, the transaction will not show up again... which is a consideration.

It only pulls details from the first document definition. It seems like there would only ever be one? I think there's multiple in the structure because this definition is what is returned by type definitions such as `gql Query { }` where you'd then have as many as defined there. I left a note in there as a TODO to still validate though.

`apollo-server/[type] [name | <anonymous>] [longest document selected path]`

#### Parsing fails: no document nor operation to go off of

In this situation, we just don't have any details to work with. Unless we want to manually grab stuff from the query (we don't), we need to make a decision on what to display.

For now, displaying `/*` to indicate could be anything :)

`apollo-server/*`

#### Batched Queries

Batched queries get executed through the plugin path separately. The current code (due to some naming API issues) is a bit clunky but the current approach is inserting "batch" when this is detected and appending all queries together with a delimiter (currently /).

`apollo-server/batch/[query 1]/[query 2]`

query 1 and query 2 are subject to the naming rules above to the right of "apollo-server".

#### Additional considerations

We could vary framework based off of apollo-server VS apollo-server-express, etc. like we would with normal web frameworks. That being said, going the plugin route, that may or may not be a pain and it is probably adequate to say "apollo-server." This would still be pretty consistent for how we'd say flag libraries built on top of "express" as "express."

### DIFF
--- a/lib/create-plugin.js
+++ b/lib/create-plugin.js
@@ -3,11 +3,126 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
- 'use strict'
+'use strict'
 
-function createPlugin() {
+function createPlugin(instrumentationApi) {
   return {
+    requestDidStart() {
+      const longestResolvedPath = {
+        depth: 0,
+        formatted: null
+      }
 
+      return {
+        executionDidStart: () => ({
+          willResolveField({info}) {
+            const pathArray = flattenToArray(info.path)
+            const formattedPath = pathArray.reverse().join('.')
+
+            if (pathArray.length > longestResolvedPath.depth) {
+              longestResolvedPath.depth = pathArray.length
+              longestResolvedPath.formatted = formattedPath
+            }
+          }
+        }),
+        willSendResponse(responseContext) {
+          // default. only used if failed to parse
+          let transactionName = '*'
+
+          if (responseContext.operation) {
+            // Resolved operation: parsed and validated, will execute resolvers
+
+            const operationType = responseContext.operation.operation
+            const operationName = responseContext.operationName || '<anonymous>'
+            const longestPath = longestResolvedPath.formatted
+
+            transactionName = `${operationType} ${operationName} ${longestPath}`
+          } else if (responseContext.document) {
+            // Failed validation, fallback to document AST
+
+            transactionName = getNameFromDocument(responseContext.document)
+          }
+
+          setTransactionName(instrumentationApi.agent, transactionName)
+        }
+      }
+    }
+  }
+}
+
+function getNameFromDocument(document) {
+  // TODO: when would there be multiple document definitions?
+  const definition = document.definitions[0]
+  const operationType = definition.operation
+  const operationName = definition.name || '<anonymous>'
+
+  const longestSelectionPath = getLongestSelectionPath(definition)
+
+  const longestPath = longestSelectionPath.join('.')
+  return `${operationType} ${operationName} ${longestPath}`
+}
+
+function flattenToArray(fieldPath) {
+  const pathArray = []
+
+  let thisPath = fieldPath
+  while (thisPath) {
+    if (typeof thisPath.key !== 'number') {
+      pathArray.push(thisPath.key)
+    }
+    thisPath = thisPath.prev
+  }
+
+  return pathArray
+}
+
+// TODO: setup so don't need special internals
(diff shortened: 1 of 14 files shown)


lib/create-plugin.js
```diff
@@ -3,11 +3,126 @@
...
+}
+
+function flattenToArray(fieldPath) {
+  const pathArray = []
+
+  let thisPath = fieldPath
+  while (thisPath) {
+    if (typeof thisPath.key !== 'number') {
+      pathArray.push(thisPath.key)
+    }
```
michaelgoin, line 72: Not sure if there's a faster way to strip the numbers here?

They seem to be representing the index of the resolver for nested resolvers but we don't really seem to have that on the document selections, and I don't think customers will find those numbers immediately obvious, so removed for consistency and clarity. Which also saves some UI space.

lib/create-plugin.js
```diff
@@ -3,11 +3,126 @@
...
+  transaction.forceName = transaction.forceName.replace(framework, `${framework}/batch`)
+  const batchName = `${transaction.forceName}/${name}`
+  transaction.forceName = batchName
+}
+
+/**
+ * Returns the longest path as an array of parts
+ * from an apollo-server document definition.
+ */
+function getLongestSelectionPath(definition) {
```
michaelgoin, line 101: I somewhat wonder if graphql (or apollo) already has a built-in function to do similar.

lib/create-plugin.js
```diff
@@ -3,11 +3,126 @@
...
+  const operationType = definition.operation
+  const operationName = definition.name || '<anonymous>'
+
+  const longestSelectionPath = getLongestSelectionPath(definition)
+
+  const longestPath = longestSelectionPath.join('.')
+  return `${operationType} ${operationName} ${longestPath}`
+}
+
+function flattenToArray(fieldPath) {
```
carlonewrelic, line 65: minor: but there are helper functions for flattening path and other things off the 'graphql' module which might be available for the plugin.
https://github.com/apollographql/apollo-server/blob/60d974240769aac3e1c75ff968657b2d3e60e9b6/packages/apollo-tracing/src/index.ts#L3

michaelgoin: ah cool. although, we might want to avoid relying on being able to import graphql / having a dependency there for something that could change. at least for something this simple. that being said, there might be other things it has we want to use which would change the consideration.

carlonewrelic (approved):
Looks great 👍 
Added a few minor comments.  Great work on testing up the tests!

lib/create-plugin.js
```diff
@@ -3,11 +3,126 @@
...
+}
+
+/**
+ * Returns the longest path as an array of parts
+ * from an apollo-server document definition.
+ */
+function getLongestSelectionPath(definition) {
+  let longestPath = null
+
+  definition.selectionSet.selections.forEach((selection) => {
```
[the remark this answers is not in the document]
michaelgoin: unsure. don't know what case validation would pass but there'd be no fields selected. something to consider as we iterate.

PR #22, opened 2020-09-12
Adds Errors from didEncounterErrors Plugin Lifecycle Event

astormnewrelic:
This PR add an Apollo plugin hook to capture errors reported by the plugin.

### DIFF
--- a/lib/create-plugin.js
+++ b/lib/create-plugin.js
@@ -4,6 +4,8 @@
  */
 
 'use strict'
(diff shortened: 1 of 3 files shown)


carlonewrelic (approved):
👍 Looks great!  I see that we are capturing errors with the didEncounterErrors hook.  The logic in the isValidRequest context was a little difficult to grok at first glance but it makes sense.

Issue #13, opened 2020-09-01
Should capture segments/spans/metrics to represent resolvers/fields


michaelgoin:
[NOTE]: # ( ^^ Provide a general summary of the request in the title above. ^^ )

## Summary

* Still some TBD here including naming.
* Nested operations (such as datastore or external calls) should be captured/nested appropriately under these constructs.

PR #23, opened 2020-09-16
Add segments

michaelgoin:
* Adds initial segment/span capture for operation and field resolution.

Closes: https://github.com/newrelic/newrelic-node-apollo-server-plugin/issues/13

Naming up for discussion/iteration but this is a starting point.

Individual requests/operations that go through the plugin model get a span named `[operation type] [operation name]` so something like `query myQuery` or `query <anonymous>`.

If there are no details to go off of, it is named `<unknown operation>`. Very open to better placeholder names, just didn't want to get bogged down getting something in place.

Resolution of field names is named after the path of the resolver. Since resolvers are executed serially (first resolver fully completes before second/nested resolver executes), the hierarchy is not obvious if you just use the field name. Using the full path makes it clear. These are currently named `resolve: <path>` so something like `resolve: libraries.books.authorname`. Again, open to however we want to evolve these names.  

---

There are some interesting situations this PR does not address.

One of which is nested resolvers may be invoked per item returned form a parent resolver. As such, we may get a ton of segments/spans. For spans, that could mean heavy sampling depending on the application. We'll want to give that some thought. My plan is to have us tackle that separate from this PR.

We also seem to have some odd behavior when an uninstrumented async method is introduced to a resolver that doesn't break anything but results in some changed naming (some middleware get //graphql appended). Will also look to have us tackle that separately.


.eslintrc
```diff
@@ -41,7 +41,11 @@
     "padded-blocks": ["error", "never"],
     "radix": "error",
     "semi": ["error", "never"],
-    "space-before-function-paren": ["error", "never"],
+    "space-before-function-paren": ["error", {
```
michaelgoin, line 44: this makes it so we can use proper formatting of `async () => {}`. otherwise, the rule forces `async() => {}` which doesn't align with how we usually do things such as `async function() {}`.

lib/create-plugin.js
```diff
@@ -10,6 +10,16 @@ const errorHelper = new ErrorHelper
 function createPlugin(instrumentationApi) {
   return {
     requestDidStart() {
+      const requestParent = instrumentationApi.getActiveSegment()
```
michaelgoin, line 13: Grabbing and passing this in is technically not requires (createSegment will do that) but I like the explicitness and we may find we want it for other uses anyways.

tests/versioned/apollo-server-express/segments.test.js.js
```diff
@@ -0,0 +1,79 @@
...
+  t.afterEach((done) => {
+    expressServer && expressServer.close()
+    server && server.stop()
+
+    helper.unload()
+    server = null
+    serverUrl = null
+    helper = null
+
+    clearCachedModules(['express', 'apollo-server-express'], () => {
```
michaelgoin, line 64: Clearing modules is critical between test runs to get consistent results. Otherwise, express won't necessarily get instrumented, etc.

tests/versioned/apollo-server-express/segments.test.js.js
```diff
@@ -0,0 +1,79 @@
...
+    const createPlugin = require('../../../lib/create-plugin')
+    const nrApi = helper.getAgentApi()
+
+    // TODO: eventually use proper function for instrumenting and not .shim
+    const plugin = createPlugin(nrApi.shim)
+
+    const express = require('express')
+
+    // Do after instrumentation to ensure express isn't loaded too soon.
+    const { ApolloServer, gql } = require('apollo-server-express')
```
carlonewrelic, line 36: Nice test for the apollo-server-express middleware

tests/versioned/apollo-server-express/segments.test.js.js
carlonewrelic: til

tests/versioned/apollo-server/transaction-naming.test.js
```diff
@@ -51,8 +52,19 @@ tap.test('apollo-server', (t) => {
...
-    done()
+    clearCachedModules(['express', 'apollo-server'], () => {
+      done()
+    })
   })
 
   createTransactionTests(t)
 })
+
+function clearCachedModules(modules, callback) {
```
carlonewrelic, line 63: Move this func to a utils file since its used in a couple places now?

tests/versioned/express-segments-tests.js
```diff
@@ -0,0 +1,661 @@
...
+    const query = `query ${expectedName} {
+      libraries {
+        books {
+          title
+          isbn
+        }
+      }
+    }`
+
+    // .isbn is the same length but title will be first so that path should be used
```
carlonewrelic, line 204: .isbn?

tests/versioned/express-segments-tests.js
```diff
@@ -0,0 +1,661 @@
...
+
+    executeQuery(serverUrl, invalidQuery, (err, result) => {
+      t.error(err)
+
+      t.ok(result)
+      t.ok(result.errors)
+      t.equal(result.errors.length, 1) // should have one parsing error
+
+      const [parseError] = result.errors
+      t.equal(parseError.extensions.code, 'GRAPHQL_PARSE_FAILED')
```
carlonewrelic, line 586: Should we have another test to check that parseError.extensions.code exists to protect against api breaking?  I'm assuming `extensions.code` is part of Apollo.

carlonewrelic (approved):
This looks good to me and great job with the really thorough tests and the added test harnesses.  I just added a few minor comments.  👍

tests/versioned/apollo-server/transaction-naming.test.js
michaelgoin: If you move the function, the resolve path will be different and it won't clear out what you are trying to clear out. There's probably a way to do generically but I'm unsure off the top of my head.

tests/versioned/express-segments-tests.js
michaelgoin: one of the queried fields... so there will be two paths of equal length [*.title, *.isbn]

tests/versioned/apollo-server/transaction-naming.test.js
carlonewrelic: Ah ok, 👍

tests/versioned/express-segments-tests.js
michaelgoin: hmm. I don't think we are doing anything related to errors that I'd expect to change fundamentals.

Perhaps? Although, I guess this sort of covers that anyways.

carlonewrelic: oh word, yeah it is covered on line 585.  please disregard :)

Issue #18, opened 2020-09-09
Return noop plugin when agent disabled


michaelgoin:
[NOTE]: # ( ^^ Provide a general summary of the request in the title above. ^^ )

## Summary

While things should be fine with the agent disabled, it may still do unnecessary work. If the agent has been disabled via configuration, we should just try to noop the plugin.

PR #24, opened 2020-09-16
Returns a noop plugin (empty object) when the agent is disabled.

michaelgoin:
Closes: https://github.com/newrelic/newrelic-node-apollo-server-plugin/issues/18

An empty object seems to work fine. Null/undefined does not.

This is currently assuming a disabled agent will have a no-op API that will not have a .shim and eventually might likely return an empty instrumentation API. We may tweak that to return a no-op of its own, so we'll have to modify as the design evolves.


carlo-808 (approved):
👍 
I see the plugin returning noop with disabled agent, integration test for new functionality and similar apollo-server versioned test.

Issue #25, opened 2020-09-17
Add versioned test for hapi


carlo-808:
Add versioned test for apollo-hapi-server

PR #32, opened 2020-09-24
Add apollo-server-hapi versioned tests

carlo-808:
- Adds apollo-server-hapi versioned test
- fixes typo server.tests.js.js file name

Closes: #25


tests/versioned/apollo-server-hapi/segments.tests.js
```diff
@@ -0,0 +1,85 @@
...
+    server = null
+    serverUrl = null
+    helper = null
+
+    clearCachedModules(['@hapi/hapi', 'apollo-server-hapi'], () => {
+      done()
+    })
+  })
+
+  createSegmentsTests(t)
```
michaelgoin, line 75: Not sure the two-file approach is really appropriate here. In the other cases, there's a shared file reused between multiple tests (two different projects that use express in different ways, for example). In this case, there's only one usage so it just sort of obfuscates things / makes them slightly more complicated than necessary.

Tis minor though, don't have to resolve with this PR.

tests/versioned/apollo-server-hapi/hapi-segments-tests.js
```diff
@@ -0,0 +1,621 @@
...
+    const query = `query {
+      hello
+    }`
+
+    helper.agent.on('transactionFinished', (transaction) => {
+      const operationPart = `query ${ANON_PLACEHOLDER}`
+      const expectedSegments = [{
+        name: `WebTransaction/apollo-server/${operationPart} hello`,
+        children: [{
+          name: 'Nodejs/Middleware/Hapi/handler//gql',
```
michaelgoin, line 32: interesting the `//gql` is in there. prob related to where with express i saw adding an async function to the middle of resolver execution resulted in `//graphql` on a pile of middleware.

michaelgoin (approved):
Looks good. One minor item called out around test structure.

tests/versioned/apollo-server-hapi/segments.tests.js
carlo-808: Great point!  I'll fix that quick.

tests/versioned/apollo-server-hapi/segments.tests.js
```diff
@@ -0,0 +1,684 @@
...
+    serverUrl = null
+    helper = null
+
+    clearCachedModules(['@hapi/hapi', 'apollo-server-hapi'], () => {
+      done()
+    })
+  })
+
+  t.test('anonymous query, single level', (t) => {
+    const { helper, serverUrl } = t.context
```
michaelgoin, line 79: As the linter is calling out, you have some unnecessary redundancy now.

You could keep the context style, I think you could grab off of an additional 't' param to afterEach() but it would probably be easier to just remove the context interaction for these tests as the items are available globally.

carlo-808: I went with removing the context interaction.

michaelgoin (approved):
👍

Issue #33, opened 2020-09-24
Invalid named query ends up with broken name


michaelgoin:
[NOTE]: # ( ^^ Provide a general summary of the issue in the title above. ^^ )

## Description

The document.name field is actually a construct with a value. Currently, it assigns it directly and ends up with '[object Object]'.

Was mid change and didn't want to munge things so capturing this as an issue but the fix is super quick/simple. Make sure to add a test case.

PR #35, opened 2020-09-28
Fixes operation name extraction when validation fails

michaelgoin:
For named queries, the operation name extraction was treating the object as a string and thus ending up with '[object Object]' instead of the proper name.

This adds the missing test case and resolves that. Also adds a file to the apollo-server versioned test runs that was missed earlier.

Fixes: https://github.com/newrelic/newrelic-node-apollo-server-plugin/issues/33


nijotz:
👀

nijotz (approved):

PR #37, opened 2020-09-29
Sets appropriate context on error capture for span/error relationship.

michaelgoin:
## Details

Errors will now be assigned to either the appropriate resolver span or the operation span for all other errors.

Since the end resolve handler will invoke before the error catch-all, had to add resolve errors at that point and then flag them as not to double-add. Its an awkward dance, but having a catch-all is kinda nice to not have to litter handlers everywhere and potentially miss a case. If it ever turns out there could be a ton of noticed errors... we might want to rethink the current implementation as it iterates an array (that should be very small).

### DIFF
--- a/lib/create-plugin.js
+++ b/lib/create-plugin.js
@@ -4,8 +4,11 @@
  */
 
 'use strict'
+
 const ErrorHelper = require('./error-helper.js')
-const errorHelper = new ErrorHelper
+const errorHelper = new ErrorHelper()
+
+const NOTICED_ERRORS = ErrorHelper.NOTICED_ERRORS
 
 function createPlugin(instrumentationApi) {
   if (!instrumentationApi) {
@@ -13,7 +16,7 @@ function createPlugin(instrumentationApi) {
   }
 
   return {
-    requestDidStart() {
+    requestDidStart(requestContext) {
       const requestParent = instrumentationApi.getActiveSegment()
 
       // We do not set to active here as batched queries will hit this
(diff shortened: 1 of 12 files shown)


lib/create-plugin.js
```diff
@@ -62,7 +70,16 @@ function createPlugin(instrumentationApi) {
                 longestResolvedPath.formatted = formattedPath
               }
 
-              return () => {
+              return (error) => {
+                if (error) {
+                  // This handler is invoked prior to didEncounterErrors
+                  // which means we need to handle the error now to capture
+                  // in context of the appropriate span.
```
nijotz, line 77: Oh, ha, this comment kind of answers my question from above. But if the didEncounterErrors code does the work to getting the appropriate context, why is this necessary? Also, this seems to push to NOTICED_ERRORS on the request context, but the didEncounterErrors code path doesn't. Does the error get sent here first, and then to didEncounterErrors, and so that code path checks for duplicates?

lib/error-helper.js
```diff
@@ -19,12 +22,26 @@ class ErrorHelper {
...
     }
   }
+
+  noticeError(instrumentationApi, error) {
+    const transaction = instrumentationApi.tracer.getTransaction()
+    instrumentationApi.agent.errors.add(transaction, error)
+  }
 }
 
+function isErrorNoticed(error, requestContext) {
```
nijotz, line 37: Looks like this finds errors added by the willResolveField code path

lib/create-plugin.js
nijotz: Ya it looks like this calls noticeError directly and didEncounterErrors will check for noticed errors and then call noticeError.

lib/create-plugin.js
```diff
@@ -62,7 +70,16 @@ function createPlugin(instrumentationApi) {
                 longestResolvedPath.formatted = formattedPath
               }
 
-              return () => {
+              return (error) => {
```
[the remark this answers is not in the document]
michaelgoin: Yes, these errors will end up in both. But this end handler will be invoked prior to didEncounter thus removing the segment from context.

lib/create-plugin.js
michaelgoin: Yeah. So this captures the errors against the appropriate resolve field segment/span. Then flags as noticed so we don't re-capture with the `didEncounterErrors` handler, which would associate to the operation segment/span.

nijotz (approved):
And it looks like all the error tests are pulled out to a separate file and applied to each framework. Nice. Looks good 👍

Issue #20, opened 2020-09-09
Add attributes


michaelgoin:
[NOTE]: # ( ^^ Provide a general summary of the request in the title above. ^^ )

## Summary

**NOTE: query has been split out of this already**

It is expected some additional research/experimentation may be completed as a part of this story.

May also want to split resolver/field VS high-level attributes into separate stories.

Attribute thoughts, thus far, after experimenting. Haven’t focused on this yet.

What do we want on error events? Same as transaction events?

Transaction Event / Other? **Note: with batch queries there will actually be two or more so maybe these can't go on transaction events**
- Operation name
- Operation type
- Deepest path?

Resolver/field Span
- Error deets once we support
- Path (as a string)
- Field name
- Return type
- Apollo’s plugin has parent type… do we care/want this?

args
- args get passed to the first resolver on the context object
- Do we want to capture just on the span? On the transaction? (focus on transaction for this issue)
- It appears (but would verify) these can only be specified at the top level by a user

Variable definitions? Where?

Is source name useful / ever something different?
```js
source: Source {
  body: 'mutation AddThing {\n      addThing(name: "added thing!")\n    }',
  name: 'GraphQL request',
  locationOffset: [Object]
}
```

PR #38, opened 2020-10-01
Add segment span attributes

michaelgoin:
## Links

* Closes: https://github.com/newrelic/newrelic-node-apollo-server-plugin/issues/20

## Details

Adds an initial set of operation and field resolve level attributes.

Field args are initially set to a destination of NONE so they do not show up unless explicitly enabled via attribute config include. We'll still want to discuss with security so they are aware and confident we are doing the right thing (we are).

Attribute names subject to change.

### DIFF
--- a/lib/create-plugin.js
+++ b/lib/create-plugin.js
@@ -10,6 +10,22 @@ const errorHelper = new ErrorHelper()
 
 const NOTICED_ERRORS = ErrorHelper.NOTICED_ERRORS
 
+const ANON_PLACEHOLDER = '<anonymous>'
+const UNKNOWN_OPERATION = '<operation unknown>'
+
+const FIELD_NAME_ATTR = 'graphql.field.name'
+const RETURN_TYPE_ATTR = 'graphql.field.returnType'
+const PARENT_TYPE_ATTR = 'graphql.field.parentType'
+const FIELD_PATH_ATTR = 'graphql.field.path'
+const FIELD_ARGS_ATTR = 'graphql.field.args'
+const OPERATION_TYPE_ATTR = 'graphql.operation.type'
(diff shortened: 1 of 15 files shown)


lib/create-plugin.js
```diff
@@ -65,9 +81,24 @@ function createPlugin(instrumentationApi) {
...
+                deepestResolvedPath.depth = pathArray.length
+                deepestResolvedPath.formatted = formattedPath
+              }
+              resolverSegment.addAttribute(FIELD_PATH_ATTR, formattedPath)
+              resolverSegment.addAttribute(FIELD_NAME_ATTR, info.fieldName)
+              resolverSegment.addAttribute(RETURN_TYPE_ATTR, info.returnType.toString())
+              resolverSegment.addAttribute(PARENT_TYPE_ATTR, info.parentType.toString())
+
+
+              for (const [key, value] of Object.entries(args)) {
```
carlo-808, line 94: Do you think there's value is adding args as attribute without a 'value'?

tests/versioned/agent-testing.js
```diff
@@ -31,8 +31,25 @@ function findSpanById(agent, spanId) {
   return matchingSpan
 }
 
+function findSegmentByName(root, name) {
```
carlo-808, line 34: nice helper func!

carlo-808 (approved):
I see attributes added for resolver segments with relevant tests for apollo-server, express and hapi.  👍

lib/create-plugin.js
michaelgoin: I considered that but ultimately decided to keep it simple and follow a similar model to request params. We can always look to add later if we find there's a need.

PR #39, opened 2020-10-02
Extracts test setup/teardown to reduce duplication.

michaelgoin:
## Details

Extracts the setup for each server type into a reusable pattern. This will make some future updates I need to make soon a bit less tedious.

Tried find the balance between reducing refactor/update pain (duplication) and not being too difficult for a new dev to understand what is going on. As such, I left some duplicative boilerplate between the various setups.

There's some pieces like the check result that prob could also be moved around but this meets the initial need for now.


carlo-808 (approved):
I see the server setup is consolidated for each respective test along with an update to the hapi tests to take server url and test utils off test context.  👍

PR #40, opened 2020-10-06
Third party license notices

nijotz:

### DIFF
--- a/THIRD_PARTY_NOTICES.md
+++ b/THIRD_PARTY_NOTICES.md
@@ -1,0 +1,503 @@
+# Third Party Notices
+
+The New Relic Node Apollo Server Plugin uses source code from third party libraries which carry
+their own copyright notices and license terms. These notices are provided
+below.
+
+In the event that a required notice is missing or incorrect, please notify us
+by e-mailing [open-source@newrelic.com](mailto:open-source@newrelic.com).
+
+For any licenses that require the disclosure of source
+code, the source code can be found at [https://github.com/newrelic/newrelic-node-apollo-server-plugin/](https://github.com/newrelic/newrelic-node-apollo-server-plugin/).
+
+## Content
+
+**[dependencies](#dependencies)**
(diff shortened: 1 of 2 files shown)


THIRD_PARTY_NOTICES.md
```diff
@@ -0,0 +1,497 @@
...
+THE SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL WARRANTIES
+WITH REGARD TO THIS SOFTWARE INCLUDING ALL IMPLIED WARRANTIES OF
+MERCHANTABILITY AND FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR
+ANY SPECIAL, DIRECT, INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
+WHATSOEVER RESULTING FROM LOSS OF USE, DATA OR PROFITS, WHETHER IN AN
+ACTION OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF OR
+IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+
+```
+
```
michaelgoin, line 497: I think we need that little footer from teh other projects in here too (just remembered)...

```
-----

Portions copyright Node.js contributors. Depending on your existing libraries and package management settings,
your systems may call externally maintained libraries in addition to those listed above.
See [here](https://nodejs.org/en/docs/meta/topics/dependencies/) and [here](https://github.com/nodejs/node/blob/v4.3.1/LICENSE)
for additional details regarding externally maintained libraries and certain related licenses and notices.”
```

Technically, we have no prod deps but we had to add everywhere during the open source work so would prob just add now to be covered.

michaelgoin (changes requested):

michaelgoin (approved):

Issue #30, opened 2020-09-22
Add query attribute


michaelgoin:
[NOTE]: # ( ^^ Provide a general summary of the request in the title above. ^^ )

## Summary

Splitting query into its own effort.

The query should be captured as an attribute. This is slightly different from datastores but seems like will want it to be tied to many of the same considerations.

Certainly, sanitation is a concern. We don't want to exposing sensitive information, unless customers are explicitly comfortable.

Should we honor the datastore configuration rules, even though this isn't a datastore for raw/obfuscated/none? Or only all/nothing via attribute configuration?

This should be discussed with security and also worth chatting w/ Vince others who have already done some work/thought around GraphQL.

### Attribute Destinations

We also need to determine the right destinations... I was originally thinking transaction event and maybe error event. The tricky part there is the batch query scenario where you could have multiple queries.

Perhaps we add to the request/operation segment/span only? Or both places? Others?

PR #41, opened 2020-10-06
Capture query in operation segment attributes

carlo-808:
## Proposed Release Notes

- Captures GraphQL query in operation segment attributes as `graphql.operation.query`
- Attempts to obfuscate all argument values in raw captured query string

## Links
Closes: #30 

## Details
- Uses a regex approach to obfuscating graphql query argument values.


lib/create-plugin.js
```diff
@@ -121,6 +123,15 @@ function createPlugin(instrumentationApi) {
           // default. only used if failed to parse
           let transactionName = '*'
           let segmentName = UNKNOWN_OPERATION
+          let query = responseContext.request.query
```
carlo-808, line 126: Might want to do this work only if agent config attributes includes 'graphql.operation.query'.  Thoughts?

lib/query-utils.js
```diff
@@ -0,0 +1,25 @@
...
+const cleanQuery = (query) => {
+  const regex = /\([\sa-zA-Z0-9,\:\"\'\{\}\$\!]+\)/g
+
+  const innerRegex = /(\:)[\sa-zA-Z0-9\"\'\$\!]+([^\{]?[,\s\}\)])/g
+
+  const matches = query.match(regex)
+  
+  if (matches) {
+    matches.forEach((match) => {
+      const newArgBlock = match.replace(innerRegex, '$1' + '***' + '$2')
```
michaelgoin, line 17: Not sure if anything more standard is done as far as replacement text like `***`. It seems fine and obvious to me. On the SQL side, we do `?` but that is a little more related to how SQL is typically written.

lib/query-utils.js
```diff
@@ -0,0 +1,14 @@
+/*
+ * Copyright 2020 New Relic Corporation. All rights reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+'use strict'
+
+const cleanQuery = (query) => {
+  const regex = /(\()[\sa-zA-Z0-9,\:\"\'\{\}\$\!]+(\))/g
```
michaelgoin, line 9: do we need all the specific checks in the middle here or could we just wildcard between two parens?

lib/query-utils.js
```diff
@@ -0,0 +1,14 @@
...
+
+  return query.replace(regex, '$1' + '***' + '$2')
```
michaelgoin, line 11: instead of separating out the matches, could prob grab the whole section and just replace with '(***)' so you don't need to concat.

michaelgoin (approved):
Looks good. Some minor comments.

I feel like we still have a gap with the query and args not being on transaction events/traces and probably errors... but still unsure what to do about the batch problem.

Something to keep simmering on.

lib/query-utils.js
carlo-808: Good point.  I'll try it out, but I think I had trouble with handling newlines with a wild card.  might work with /s/S

michaelgoin (approved):
👍

Issue #31, opened 2020-09-22
Finalize transaction naming


michaelgoin:
[NOTE]: # ( ^^ Provide a general summary of the request in the title above. ^^ )

## Summary

* Ensure naming implementation adequate from a details/user perspective
* Ensure naming implementation adequate/valid from a back-end perspective
  * What considerations/best practices are there?
  * What things should actually be done via UI instead of string formatting of name?
*  Use/modify/implement fully API supported way of naming without using hands-off internals.

For example: perhaps there is a consideration where it would be better to use a delimeter (such as 'detail1/detail2') and have the UI handle the display formatting.

michaelgoin:
Decided today we are OK from a naming perspective. May consider removing deepest-path from batch queries at some point

PR #42, opened 2020-10-06
Updates transaction and segment/span naming 

michaelgoin:
## Links

* Closes: https://github.com/newrelic/newrelic-node-apollo-server-plugin/issues/31

## Details

Updates to more standard naming conventions that separate pieces by '/'.

Transactions will now be named based on the web framework, via web convention using the method/verb. We may decide to tweak that more over time.


carlo-808 (approved):
👍 
I see the transaction and segment/span naming schemes are updated and all tests updated as well.

Issue #36, opened 2020-09-29
Add a way to track plugin usage


michaelgoin:
We should have a way to see usage of the plugin and keep track of uptake.

Not currently sure the best way here. One option is to add a new (or leverage existing?) support metric and add to angler. Could be specific to plugin or maybe tracked as a part of the API improvement in issue: https://github.com/newrelic/newrelic-node-apollo-server-plugin/issues/14.

PR #43, opened 2020-10-08
Adds initial metrics

michaelgoin:
## Links

* Closes: https://github.com/newrelic/newrelic-node-apollo-server-plugin/issues/36
* Dashboard example leveraging metrics: [a long url](https://staging-one.newrelic.com/launcher/dashboards.launcher?pane=eyJuZXJkbGV0SWQiOiJkYXNoYm9hcmRzLmRhc2hib2FyZCIsImVudGl0eUlkIjoiTVRrd2ZGWkpXbnhFUVZOSVFrOUJVa1I4TVRZeU5qQTAiLCJ1c2VEZWZhdWx0VGltZVJhbmdlIjpmYWxzZSwiaXNUZW1wbGF0ZUVtcHR5IjpmYWxzZX0=&platform[$isFallbackTimeRange]=false&platform[accountId]=190)

## Details

* Operation segment names now also include the deepest path.
* Operation segments now use a final name part of '`<unknown>`' instead of '`<operation unknown>`'
  The additional 'operation' word was redundant now that 'operation' is an earlier part of the name (prior pr).
* Adds initial metrics for operations and field resolve.
  * Operation metrics match 1:1 with segment name
  * Resolve metrics use the field name **NOT** the path for comparison across all usages.
    We may eventually want to generate a path metric too? I figure it is easier to add later than remove something we don't need.
* Adds support metric to track usage/uptake of the module.

Some test files got moved around so they can be leveraged outside the versioned tests.

Support metric will need to be added to visualization tracking.


tests/integration/metrics.test.js
```diff
@@ -0,0 +1,181 @@
...
+  t.test('anonymous query, single level metrics', (t) => {
+    const { helper, serverUrl } = t.context
+
+    const query = `query {
+      hello
+    }`
+
+    helper.agent.on('transactionFinished', () => {
+      const operationPart = `query/${ANON_PLACEHOLDER}/hello`
+      t.metrics([
```
carlo-808, line 37: This must be the tap extensions from test-utils 👍

tests/integration/metrics.test.js
```diff
@@ -0,0 +1,181 @@
...
+    const invalidQuery = `query {
+      libraries {
+        books {
+          title
+          author {
+            name
+          }
+        }
+      }
+    ` // missing closing }
```
carlo-808, line 167: Nice calling out whats invalid about query in comment

lib/create-plugin.js
```diff
@@ -279,4 +283,45 @@ function getDeepestSelectionPath(definition) {
...
 }
 
+function recordResolveSegment(segment, scope) {
+  const duration = segment.getDurationInMillis()
+  const exclusive = segment.getExclusiveDurationInMillis()
+
+  const transaction = segment.transaction
+
+  const attributes = segment.getAttributes()
+  const fieldName = attributes[FIELD_NAME_ATTR]
```
carlo-808, line 293: I think this will work with an alias for a field.  minor: add a test with an alias?

carlo-808 (approved):
I see we are adding metrics for Operation and Resolver timings with some nice helper functions for tests.  👍

lib/create-plugin.js
michaelgoin: We don't really have anything considering alias' at all.

Perhaps enter a new issue to explore the impacts of aliases and add tests to the various considerations (transaction naming, segment naming, attributes, metrics).

carlo-808: Sounds good 👍

PR #46, opened 2020-10-12
Exclude scalars

michaelgoin:
## Details

Nested scalar fields (included defined as non-null) will no-longer be captured as spans/metrics by default. This means that time (if any) will be associated with the operation itself. Top level scalar fields (query resolver that returns a string for example) will still get captured.

Captured:

```
type Query {
    books: [Book] // Captured
    hello: String // Captured
}

type Book {
    title: String! // NOT CAPTURED
    author: Author! // Captured
}
```

The plugin nowr eturns a `createPlugin` function that allows passing of configuration. This is pre-bound to the instrumentation API.

Currently, the configuration will allow for re-enabling of scalar fields should that be desired via `{ captureScalars: true }`. **NOTE: we could also ship without this option and wait to see if needed?**.

Since Apollo Server will invoke any plugin passed in as a function, customers can still use the previous approach.

```js
const plugin = require(@newrelic/apollo-server-plugin)
  typeDefs,
  resolvers,
  plugins: [plugin]
});
```

OR

```js
const createPlugin = require(@newrelic/apollo-server-plugin)
const plugin = createPlugin({captureScalars: true})
  typeDefs,
  resolvers,
  plugins: [plugin]
});
```

This does not solve the issue of many nested resolvers used on list items.

### DIFF
--- a/tests/versioned/apollo-server-hapi/apollo-server-hapi-setup.js
+++ b/tests/versioned/apollo-server-hapi/apollo-server-hapi-setup.js
@@ -14,7 +14,7 @@ const { getTypeDefs, resolvers } = require('../../data-definitions')
 
 const WEB_FRAMEWORK = 'Hapi'
 
-function setupApolloServerHapiTests({suiteName, createTests}, config) {
+function setupApolloServerHapiTests({suiteName, createTests, pluginConfig}, config) {
   tap.test(`apollo-server-hapi: ${suiteName}`, (t) => {
     t.autoend()
 
@@ -30,7 +30,7 @@ function setupApolloServerHapiTests({suiteName, createTests}, config) {
       const nrApi = helper.getAgentApi()
 
       // TODO: eventually use proper function for instrumenting and not .shim
-      const plugin = createPlugin(nrApi.shim)
+      const plugin = createPlugin(nrApi.shim, pluginConfig)
 
       const Hapi = require('@hapi/hapi')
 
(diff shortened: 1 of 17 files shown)


tests/versioned/apollo-server-hapi/apollo-server-hapi-setup.js
```diff
@@ -14,7 +14,7 @@ const { getTypeDefs, resolvers } = require('../../data-definitions')
 
 const WEB_FRAMEWORK = 'Hapi'
 
-function setupApolloServerHapiTests({suiteName, createTests}, config) {
+function setupApolloServerHapiTests({suiteName, createTests, pluginConfig}, config) {
```
carlo-808, line 17: Do we need to update the koa and fastify versioned test setup as well?

michaelgoin: Yeah, those will need to be updated. They didn't exist yet when I made these changes :P

carlo-808: I can make those updates with my lambda versioned test PR.

michaelgoin: i'd be all about not having to do it but those tests will likely break on main if this is merged as the default is to now exclude scalars (except for the top level)

carlo-808 (approved):
I see that we are adding a config option to the plugin to capture scalars or not.  CaptureScalars is defaulted false if no config is provided to plugin.  Associated tests were added for filtering scalar segments.  👍

carlo-808 (approved):
I see that we are adding a config option to the plugin to capture scalars or not. CaptureScalars is defaulted false if no config is provided to plugin. Associated tests were added for filtering scalar segments. 👍

PR #48, opened 2020-10-14
Add AWS Lambda versioned tests

carlo-808:
## Proposed Release Notes

- Support for `apollo-server-lambda`

## Links
Closes: #48 
## Details
- Added versioned tests for aws lambda support.
- Updated third party manifest and notices to include `apollo-server`


.gitignore
```diff
@@ -59,3 +59,6 @@ typings/
 
 # VS Code specific configuration
 .vscode
+
+# macOs file
+.DS_Store
```
michaelgoin, line 64: minor: missing a newline at the end.

michaelgoin (commented):
Is the only difference between the lambda specific versions of tests and the existing versions (not counting segments tests which will differ) the use of the lambda queries for requests?

carlo-808:
> Is the only difference between the lambda specific versions of tests and the existing versions (not counting segments tests which will differ) the use of the lambda queries for requests?

For most of those tests, I had to update the query execution and pass around some objects on the tap context.  For the error test, I had to look at the segments rather than the spans to check for error attributes.  Something about the lambda handler execution caused no spans to be generated when the transaction finished.  That only happened in the error tests and might be due to test setup(?) but I didn't want to get bogged down there.

michaelgoin:
@carlo-808 OK. I think we probably could have setup a consistent API for making requests and swapped out passing different ones but that can be a future refactor.

Only thing that concerns me is if that span thing is an actual issue. 

Although, it seems you are not enabling distributed tracing in your errors test setup (did you remove that or is that the cause of the problem?) and also the test is not running in serverless_mode like a real lambda would.

tests/versioned/apollo-server-lambda/apollo-server-lambda-setup.js
```diff
@@ -27,7 +28,10 @@ function setupApolloServerLambdaTests({suiteName, createTests, pluginConfig}, co
     t.beforeEach(async () => {
       const { ApolloServer, gql } = require('apollo-server-lambda')
 
+      agentTesting.temporarySetEnv(t, 'NEW_RELIC_ACCOUNT_ID', 'eeeeee')
```
michaelgoin, line 31: this is using the outer `t` so could be before/outside the beforeEach and is not needed to be called multiple times that way.

if you want tied to the individual test run, a `t` param does get passed to `beforeEach`. 

in callback style, that looks like `t.beforeEach((done, t) => {})` not sure when you are using async...

michaelgoin, line 31: i might leave a comment as to "why" this is necessary, given the other configuration items are in the config file and the linkage won't be obvious to newcomers.

michaelgoin (dismissed):
Looks good. Some minor comments for consideration.

tests/versioned/apollo-server-lambda/apollo-server-lambda-setup.js
carlo-808: I think async is a copy/pasta fail.  I'll fix it.

carlo-808: Sounds good, I'll update.

michaelgoin: ah yeah, maybe from when you were doing koa. i shoulda checked for that.

so yeah, use callback style and invoke the `done()` and use the `t` and that will cleanup after each individual test. which isn't necessary, it could be before all, but since the rest of the setup/teardown is per test might as well stay consistent.

michaelgoin (approved):
👍

PR #49, opened 2020-10-19
Remove template text from readme

nijotz:
## Details
* Went over README, comparing it to the main agent README, removed template text.

### DIFF
--- a/README.md
+++ b/README.md
@@ -201,7 +201,7 @@ npm run versioned
 
 ## Support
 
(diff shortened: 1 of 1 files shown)


michaelgoin (approved):
We probably could have left the "diagnostics tools" w/o the template brackets since we have links to NRDiag, etc. but I don't think it hurts to be removed, either.

nijotz:
Oh good point, I'll throw it in there bracketless.

PR #51, opened 2020-10-20
Update readme with plugin timings recommendations

carlo-808:
## Details
Update readme with Apollo server setup recommendations when multiple plugins are in use.

Other plugins in the Apollo Server setup will affect the overall timings of a graphql request.  If the New Relic Apollo Plugin is last in the chain of plugins used, the field resolution timings will be more accurate.


nijotz (approved):
Nice work researching this! Rad addition. Maybe line-wrapping love for the old men in vim, but looks good!

nijotz (approved):
👍

PR #56, opened 2020-10-29
Only set query attribute if it exists

MightySCollins:
## Proposed Release Notes
Fixed error when no query is provided

## Links
https://www.apollographql.com/docs/apollo-server/performance/apq/

## Details
In cases when apollo client is using persisted queries the query parameter is not sent in the request. In theory you could set the query has as an attribute but I just wanted to fix the 500 error.

This fixes a 500 error we had where cleanQuery was doing a regex on the value undefined.
```
TypeError: Cannot read property 'replace' of undefined
```

### DIFF
--- a/lib/create-plugin.js
+++ b/lib/create-plugin.js
@@ -166,12 +166,14 @@ function createPlugin(instrumentationApi, config = {}) {
           let segmentName = UNKNOWN_OPERATION
           let query = responseContext.request.query
 
-          // attempt to extract arguments to strip from query
-          query = cleanQuery(query)
+          if (query) {
+            // attempt to extract arguments to strip from query
+            query = cleanQuery(query)
 
-          operationSegment.addAttribute(
-            OPERATION_QUERY_ATTR,
-            query)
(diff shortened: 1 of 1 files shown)


nijotz:
https://github.com/apollographql/apollo-server/blob/main/packages/apollo-server-types/src/index.ts#L88

Yup, looks like it is optional

carlo-808:
Great catch @MightySCollins and thank you for the quick fix!  We'll push this fix through and get a patch release out in short order.

Are you using `apollo-link-persisted-queries` to persist the queries?  I just added a bug fix issue to capture the query whether its persisted or not and if you could add any extra context to it that would be awesome!  
https://github.com/newrelic/newrelic-node-apollo-server-plugin/issues/57

Thanks!

carlo-808 (approved):
Great fix!  👍

nijotz:
Hey @MightySCollins, we just released v0.1.1 of this plugin with your fix if you want to upgrade. Thanks again for the contribution!

Issue #59, opened 2020-11-03
Cannot read property 'start' of null" prevents request


mschipperheyn:
## Description

I'm getting this error. It doesn't output in the server logs, but does show up on the client. The error completely prevents the request from processing. 

newrelic.js
```
exports.config = {
   agent_enabled: true,
   ssl: false,
   app_name: ['FNB Server'],
   license_key: 'XYZ',
   distributed_tracing: {
	enabled: false,
   },
   logging: {
     level: 'info'
   },
   allow_all_headers: true,
   attributes: {
      		exclude: [
			'request.headers.cookie',
			'request.headers.authorization',
			'request.headers.proxyAuthorization',
			'request.headers.setCookie*',
			'request.headers.x*',
			'response.headers.cookie',
			'response.headers.authorization',
			'response.headers.proxyAuthorization',
			'response.headers.setCookie*',
			'response.headers.x*',
		],
   }
}
```
## Expected Behavior

- In case of process errors, it should not prevent the request from completing
- Errors in processing should be logged.

## Your Environment

[TIP]:  # ( Include as many relevant details about your environment as possible. )

* Chrome
* newrelic: 6.14.0
* @newrelic/apollo-server-plugin: 0.1.1
* apollo-server-express: 2.18.2
* Node version: 12.18.2
* Mac OS

nijotz:
Hi Marc, would you mind pasting the full stacktrace that you received to help us debug?

michaelgoin:
Doing a quick scan, I'm guessing this is with the `segment.start()` calls? Seems like maybe there's some transaction state loss (which may want to work through via support ticket).

That being said, you are right in that we don't want the chance to throw an error and error out the GraphQL request.

mschipperheyn:
Hi guys, let me see here. 
The stack is:
```
TypeError: Cannot read property 'start' of null"
    at Object.requestDidStart (/Users/mschipperheyn/projects/ology-rnw/packages/server/Users/mschipperheyn/projects/ology-rnw/node_modules/@newrelic/apollo-server-plugin/lib/create-plugin.js:82:1)"
    at initializeRequestListenerDispatcher (/Users/mschipperheyn/projects/ology-rnw/packages/server/build/webpack:/Users/mschipperheyn/projects/ology-rnw/node_modules/apollo-server-core/dist/requestPipeline.js:305:1)"
[...]

PR #62, opened 2020-11-09
Fixes bug improperly handling state loss prior to plugin execution.

michaelgoin:
## Proposed Release Notes

* Fixed bug that would cause errors if transaction state loss occurred prior to plugin execution.

## Links

* Fixes: https://github.com/newrelic/newrelic-node-apollo-server-plugin/issues/59

## Details

This will fix the error part of the attached issue. The customer will still have state loss issues (unsure the cause).

One thing for us to consider is if we want to wrap some (or all) of these handlers in `try/catch` blocks to prevent accidental future bugs from erroring out customer queries. There can be some complication there with our usage of closures and any potential overhead. Chose not to tackle that with this effort.

### DIFF
--- a/tests/integration/state-loss.test.js
+++ b/tests/integration/state-loss.test.js
@@ -1,0 +1,65 @@
+/*
+ * Copyright 2020 New Relic Corporation. All rights reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+'use strict'
+
+const { executeQuery } = require('../test-client')
+const { setupEnvConfig } = require('../agent-testing')
+
+const { setupApolloServerTests } = require('../apollo-server-setup')
+
+setupApolloServerTests({
+  suiteName: 'State Loss',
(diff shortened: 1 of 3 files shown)


tests/integration/state-loss.test.js
```diff
@@ -0,0 +1,65 @@
...
+      t.end()
+    })
+  })
+}
+
+function createStateLossPlugin(instrumentationApi) {
+  return {
+    requestDidStart() {
+      // Setting active segment to null to mimic state loss
+      instrumentationApi.setActiveSegment(null)
```
nijotz, line 62: neat

nijotz (approved):
Looks good. Why not check `currentSeg`? Because of the active segment check at 74?

michaelgoin:
@nijotz currentSeg just gets restored for the current state upon resolve end... does not impact segment creation. as long as we have the operation segment, we are fine to create a resolve segment.

Issue #67, opened 2021-01-29
Cannot read property 'nameState' of null


frangrolemund:
## Description

After performing the NewRelic Apollo integration I've started noticing the following in our logs.  This doesn't mean a whole lot but did notice this reported in the [koa integration](https://github.com/newrelic/node-newrelic-koa/issues/32) back in June - wondering if maybe it is related here.  Apologies for the poor format:

`2021-01-29T10:44:44.636Z [WARN] Cannot read property 'nameState' of null Context: {"error":{"message":"Cannot read property 'nameState' of null","extensions":{"code":"INTERNAL_SERVER_ERROR","exception":{"stacktrace":["TypeError: Cannot read property 'nameState' of null"," at setTransactionName (/usr/cocina/node_modules/@newrelic/apollo-server-plugin/lib/create-plugin.js:314:33)"," at Object.willSendResponse (/usr/cocina/node_modules/@newrelic/apollo-server-plugin/lib/create-plugin.js:224:11)"," at /usr/cocina/node_modules/apollo-server-core/dist/utils/dispatcher.js:21:31"," at Array.map (<anonymous>)"," at Dispatcher.callTargets (/usr/cocina/node_modules/apollo-server-core/dist/utils/dispatcher.js:18:24)"," at Dispatcher.<anonymous> (/usr/cocina/node_modules/apollo-server-core/dist/utils/dispatcher.js:27:43)"," at Generator.next (<anonymous>)"," at /usr/cocina/node_modules/apollo-server-core/dist/utils/dispatcher.js:8:71"," at new Promise (<anonymous>)"," at __awaiter (/usr/cocina/node_modules/apollo-server-core/dist/utils/dispatcher.js:4:12)"]}},"name":"TypeError"}}`

## Expected Behavior

I'd like to not have this error appear in the logs either by changing my integration somehow or updating the package if necessary.

## Steps to Reproduce
Not clear what triggers this error.

## Your Environment

Node v12.14.1
Apollo Server Express 2.17.0

## Additional context

We're also instrumented with AWS Xray as well.

michaelgoin:
Interesting. Really surprised there would be state loss introduced mid-plugin execution.

Do you by chance have any other custom or 3rd-party plugins in use?

We can handle this symptom, though. We can grab the transaction off of the operation segment.

Thank you for letting us know and sorry for the inconvenience!

realproven:
Thanks for looking at this, much appreciated!

[...]

PR #69, opened 2021-02-23
Fixes issues setting transaction name when state loss occurs.

michaelgoin:
## Proposed Release Notes

* Fixed issue where state loss introduced through another module could result in the plugin causing a query to error while trying to set the transaction name.

## Links

* Fixes: https://github.com/newrelic/newrelic-node-apollo-server-plugin/issues/67

## Details

This does not solve the underlying issue of state loss, which is potentially from the usage of AWS XRay in the linked issue. That is a problem likely outside this plugin.

### DIFF
--- a/tests/integration/state-loss.test.js
+++ b/tests/integration/state-loss.test.js
@@ -27,7 +27,7 @@ function createStateLossTests(t) {
     done()
   })
 
-  t.test('should should not error when state loss prior to query', (t) => {
+  t.test('should not error when state loss prior to query', (t) => {
     stateLossTester.triggerOnRequestDidStart()
 
     const { serverUrl } = t.context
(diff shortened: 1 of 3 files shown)


tests/integration/state-loss.test.js
```diff
@@ -53,13 +63,41 @@ function createStateLossTests(t) {
...
-}
 
-function createStateLossPlugin(instrumentationApi) {
-  return {
-    requestDidStart() {
-      // Setting active segment to null to mimic state loss
-      instrumentationApi.setActiveSegment(null)
-    }
-  }
+  t.test('should should not error when state loss just before sending response', (t) => {
```
carlo-808, line 67: minor: 2 shoulds :)

carlo-808 (changes requested):

tests/integration/state-loss.test.js
michaelgoin: I'M SERIOUS CARLO IT SHOULD

carlo-808 (approved):
👍

PR #73, opened 2021-04-26
fix: package.main and upgrade newrelic

wbruno:
## Proposed Release Notes
Update newrelic dependency and fix package.main.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,8 +1,7644 @@
 {
   "name": "@newrelic/apollo-server-plugin",
   "version": "0.1.3",
-  "lockfileVersion": 1,
+  "lockfileVersion": 2,
   "requires": true,
+  "packages": {
+    "": {
+      "name": "@newrelic/apollo-server-plugin",
+      "version": "0.1.3",
+      "license": "Apache-2.0",
+      "devDependencies": {
+        "@newrelic/test-utilities": "^4.1.2",
+        "apollo-server": "^2.18.2",
+        "eslint": "^7.8.1",
+        "newrelic": "^7.3.1",
+        "tap": "^14.10.8"
+      },
+      "engines": {
+        "node": ">=10.0.0"
+      },
+      "peerDependencies": {
+        "newrelic": ">=7.3.1"
+      }
+    },
+    "node_modules/@apollo/protobufjs": {
+      "version": "1.0.5",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -2,7 +2,7 @@
   "name": "@newrelic/apollo-server-plugin",
   "version": "0.1.3",
   "description": "Apollo Server plugin that adds New Relic Node.js agent instrumentation.",
-  "main": "/src/index.js",
+  "main": "index.js",
```
michaelgoin, line 5: 👍 Thanks!

package.json
```diff
@@ -31,11 +31,10 @@
     "@newrelic/test-utilities": "^4.1.2",
     "apollo-server": "^2.18.2",
     "eslint": "^7.8.1",
-    "newrelic": "^6.13.0",
+    "newrelic": "^7.3.1",
```
michaelgoin, line 34: This works fine but could also be `>=6.13.0`.

package.json
```diff
@@ -31,11 +31,10 @@
...
     "tap": "^14.10.8"
   },
   "peerDependencies": {
-    "newrelic": ">=6.13.0"
-  },
-  "dependencies": {}
+    "newrelic": ">=7.3.1"
```
michaelgoin, line 38: We currently are preferring this to be a peer dependency. We only allow for one version of the agent to be loaded at a time and will use (regardless) what the application is currently setup to use. While we do have a safety mechanism in place for if this is set as a standard dependency, it will log an unnecessary warning for trying to load the agent more than once.

I would revert this part of the change.

package-lock.json
```diff
@@ -1,8 +1,7644 @@
 {
   "name": "@newrelic/apollo-server-plugin",
   "version": "0.1.3",
-  "lockfileVersion": 1,
+  "lockfileVersion": 2,
```
michaelgoin, line 4: Interested in the change here...

What version of npm are you using?

michaelgoin (changes requested):
Thank you for the contribution!

I left some specific comments in the related code areas for consideration.

package.json
wbruno: got it! will do

package-lock.json
wbruno:
``` sh
$ npm -v
7.10.0
```

i will revert it

package.json
wbruno: okay!

PR #74, opened 2021-05-04
fix: package.json main field

dnalborczyk:
## Proposed Release Notes

the `main` field in package.json is pointing to a wrong location.

**node.js v16+** prints a warning:

```
DeprecationWarning: Invalid 'main' field in '/node_modules/@newrelic/apollo-server-plugin/package.json' of '/src/index.js'. Please either fix that or report it to the module author
(Use `node --trace-deprecation ...` to show where the warning was created)
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,7 +2,7 @@
   "name": "@newrelic/apollo-server-plugin",
   "version": "0.1.3",
   "description": "Apollo Server plugin that adds New Relic Node.js agent instrumentation.",
-  "main": "/src/index.js",
+  "main": "./index.js",
   "scripts": {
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -2,7 +2,7 @@
   "name": "@newrelic/apollo-server-plugin",
   "version": "0.1.3",
   "description": "Apollo Server plugin that adds New Relic Node.js agent instrumentation.",
-  "main": "/src/index.js",
+  "main": "./index.js",
```
michaelgoin, line 5: Could also just have `"main:": "index.js"` but this seems to be valid.

michaelgoin (approved):
Thanks!

PR #79, opened 2021-05-12
Runs fastify versioned tests aonly against latest fastify version.

michaelgoin:
## Proposed Release Notes

* Pins fastify to a single version for apollo-server-fastify versioned testing.

## Details

We don't really need to be testing a big permutation of fastify itself, as our instrumentation is specific to the plugin. If something breaks with a fastify version, it is most likely going to be with the plugin framework or apollo-server-fastify. Drops permutations down to ~44 and reduces execution time by about 2/3.

Ideally, would prob use the latest version each run but I couldn't determine an obvious way to do that with the versioned framework right now and didn't want to invest a ton of time.

### DIFF
--- a/tests/versioned/apollo-server-fastify/package.json
+++ b/tests/versioned/apollo-server-fastify/package.json
@@ -9,7 +9,7 @@
       },
       "dependencies": {
         "apollo-server-fastify": ">=2.14.0",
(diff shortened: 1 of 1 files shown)


carlo-808 (approved):
👍

PR #80, opened 2021-05-12
Bump versioned test runner and prep for Node 16

michaelgoin:
## Proposed Release Notes

* Bumped @newrelic/test-utilities to ^5.1.0.

* Added npm scripts and GitHub actions plumbing to run versioned test runner with appropriate flags for npm v7 and npm v6.

  Node 16 ships with npm v7 which has several different behaviors than npm v6. Testing found we need to ensure all packages are installed for each versioned test permutation which is handled via the `--all` flag in the new version of the versioned test runner.

## Links

* Resolves some of the issues found in: https://github.com/newrelic/newrelic-node-apollo-server-plugin/pull/75

## Details

Wired up handling for Node 16 but not running against 16 as there are other issues to be resolved. This takes Bob's update to the versioned runner and hooks it up to help move the other PR(s) forward.

### DIFF
--- a/.github/workflows/ci-workflow.yml
+++ b/.github/workflows/ci-workflow.yml
@@ -72,5 +72,9 @@ jobs:
         node-version: ${{ matrix.node-version }}
     - name: Install Dependencies
       run: npm ci
-    - name: Run Versioned Tests
-      run: npm run versioned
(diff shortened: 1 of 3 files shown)


carlo-808 (approved):
👍  Updated to use latest test-utilities module and npm 7 versioned tests for Node 16.

Issue #87, opened 2021-06-16
Fix segment nesting for federated gateway servers


michaelgoin:
For federated servers, the external segments that should be children of the operation are nested at the same level under the middleware. This is because the operation segment is not set to active.

Because we are keeping track of resolvers and manually associating with the operation, we don’t have a problem on normal server instrumentations. 

We do not currently set the operation to active because it causes problems with batch queries. The operations for batch queries start back-to-back, then resolvers execute asynchronously and first to end is the first to hit ‘willSendResponse’ if I remember right. As such, there’s some consideration here as not to bleed transaction state.

This may or may not be easy, given the different use cases but the proper nesting is important for context in the federated case.

michaelgoin:
what about batch queries to federated gateway? may not be able to solve this in a good way via the plugin.

PR #95, opened 2021-06-25
Fixes segment nesting for Apollo Federation Gateway operations.

michaelgoin:
## Proposed Release Notes

* Fixed segment nesting for Apollo Federation gateway operations.

## Links

* Fixes: https://github.com/newrelic/newrelic-node-apollo-server-plugin/issues/87

## Details

* Tests Node >=12.13.0 per current requirements of @apollo/federation.

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,6 +10,7 @@
     "test": "npm run unit && npm run integration && npm run versioned",
     "unit": "tap --test-regex='(\\/|^tests\\/unit\\/.*\\.test\\.js)$' --no-coverage",
     "versioned": "npm run versioned:npm6",
+    "versioned:folder": "versioned-tests --minor -i 2",
     "versioned:npm6": "versioned-tests --minor -i 2 'tests/versioned/*'",
     "versioned:npm7": "versioned-tests --minor --all -i 2 'tests/versioned/*'"
   },
--- a/tests/versioned/apollo-federation/federated-gateway-server-setup.js
+++ b/tests/versioned/apollo-federation/federated-gateway-server-setup.js
@@ -1,0 +1,191 @@
+/*
+ * Copyright 2021 New Relic Corporation. All rights reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+'use strict'
+
+const tap = require('tap')
+
+const utils = require('@newrelic/test-utilities')
+utils.assert.extendTap(tap)
+
+const federatedData = require('./federated-data-definitions')
+
+const WEB_FRAMEWORK = 'Expressjs'
+
+function setupFederatedGatewayServerTests(options, agentConfig) {
+  const {
+    suiteName,
+    createTests,
+    pluginConfig
+  } = options
+
+  tap.test(`apollo-federation: ${suiteName}`, (t) => {
+    t.autoend()
+
+    let gatewayServer = null
+    let libraryServer = null
+    let bookServer = null
+    let magazineServer = null
+
+    let helper = null
+
+    t.beforeEach(async () => {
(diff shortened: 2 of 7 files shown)


package.json
```diff
@@ -10,6 +10,7 @@
     "test": "npm run unit && npm run integration && npm run versioned",
     "unit": "tap --test-regex='(\\/|^tests\\/unit\\/.*\\.test\\.js)$' --no-coverage",
     "versioned": "npm run versioned:npm6",
+    "versioned:folder": "versioned-tests --minor -i 2",
```
michaelgoin, line 13: This just made it easier to run only the versioned tests I wanted locally via `npm run versioned:folder ./tests/versioned/apollo-federation`

tests/versioned/apollo-federation/federated-gateway-server-setup.js
```diff
@@ -0,0 +1,191 @@
...
+}
+
+function createIgnoreTransactionPlugin(nrApi) {
+  return {
+    requestDidStart() {
+      const transactionHandle = nrApi.getTransaction()
+      transactionHandle.ignore()
+    }
+  }
+}
```
michaelgoin, line 167: I didn't want to overly complicate things by spinning off multiple processes, at least as of yet. Particularly since we may eventually want to validate through each hop at some point.

That does make things a little awkward, though, because even without using our plugin transactions will be created via our http and express instrumentation. There's really no way to avoid this without a separate process.

May decide to go that route at some point but for now I am ignoring transactions for the sub graph services to avoid creating the typical data and then checking for ignored on transaction finished in the tests to avoid apply assertions.

tests/versioned/apollo-federation/segments.test.js
```diff
@@ -0,0 +1,192 @@
...
+          issue,
+          title
+        }
+      }
+    }`
+
+    helper.agent.on('transactionFinished', (transaction) => {
+      if (transaction.forceIgnore) {
+        return
+      }
```
michaelgoin, line 51: This is ignoring the sub-graph transactions. See the related comment in the gateway server setup file.

tests/versioned/apollo-federation/segments.test.js
```diff
@@ -0,0 +1,192 @@
...
+                  { name: libraryExternal },
+                  { name: bookExternal }
+                ]
+              },
+              {
+                name: `${OPERATION_PREFIX}/${operationPart2}`,
+                children: [
+                  { name: libraryExternal },
+                  { name: magazineExternal }
+                ]
```
michaelgoin, line 150: Verifies the correct external segments are nested under each batch query.

tests/versioned/apollo-federation/segments.test.js
```diff
@@ -0,0 +1,192 @@
...
+  t.ok(result)
+
+  if (result.errors) {
+    result.errors.forEach((error) => {
+      t.error(error)
+    })
+  }
+
+  setImmediate(callback)
+}
```
michaelgoin, line 185: This is getting enough uses now it probably could get extracted into its own module for reuse. I didn't tackle with this PR but likely will do so (or someone else can) later.

bizob2828 (changes requested):
A few questsions/comments.  I was able to verify the nesting looks good          

screenshot 2021-06-29 at 3 08 50 PM

It looks like since you opened PR a new version of one of the deps was released and it broke CI.  I'm goin to kick this back to you for more triaging

bizob2828 (approved):
Good work, this will make the DX much better when using an ApolloGateway and properly nesting federated queries under the apollo server request

Issue #111, opened 2021-07-12
Name transactions by deepest unique path


michaelgoin:
The current naming has a few issues:
* It can sometimes be arbitrary (choosing first of two identical length paths)
* It can hide commonalities where slowness may occur

By stopping naming at the point by which multiple fields are selected, we can avoid arbitrary decisions and accidental obfuscation of where slowness may occur.

For each selection set, add to the path name if only one (non-filtered) field has been selected. If multiple (non-filtered) fields are in the selection set, accept the current path for the name.

By default, we will ignore `id` and `__typename` fields which are commonly a part of sub graphs and would make naming unnecessarily vague.

TBD:
  * Naming if immediately querying multiple resolvers at the top-level? NerdGraph does not include a path in this case as anything chosen would be misleading. Seems like we probably could/should start here and recommend good naming of the queries for these situations (understanding customers don't always control the incoming queries).
  * Should we start with user-configurable filter list for naming or wait til there is a clear need? 

Quick prototype: https://github.com/newrelic/newrelic-node-apollo-server-plugin/compare/main...michaelgoin:unique-naming-field-filter-proto?expand=1

michaelgoin:
We have to decide to go with this naming improvement over https://github.com/newrelic/newrelic-node-apollo-server-plugin/issues/99 due to the complexity of the solution. This will be a general improvement and for most use cases as good as the other proposal. Unfortunately, the scalar filtering proved rather complex on initial prototyping.

PR #113, opened 2021-07-13
Deepest unique naming

michaelgoin:
## Proposed Release Notes

* **BREAKING**: modifies transaction and operation segment/span naming to use deepest *unique* path instead of first deepest path.

  The deepest unique path is the deepest path in the selection set of a query where only one field was selected at each level. 'id' and '__typename' fields are automatically excluded from this decision. For example: `query { libraries { branch booksInStock magazinesInStock }}` will use the path name 'libraries' and  `query { libraries { branch __typename id }}` will use the path name 'libraries.branch'. For more information, see the latest [transaction naming documentation](https://github.com/newrelic/newrelic-node-apollo-server-plugin/blob/main/docs/transactions.md).
  
  This may result in naming changes for your existing GraphQL queries. Any charts, etc. where you are querying by transaction name will need to be updated upon upgrading.

* **BREAKING**: removed the deepest path attribute 'graphql.operation.deepestPath' from operation segments/spans.

## Links

* Closes: https://github.com/newrelic/newrelic-node-apollo-server-plugin/issues/111

## Details

* ~Transaction documentation still needs to be updated.~ 

* Apologies for all the test noise. Needs some further thought but we should consider how we may simplify the tests. Extracting queries + expected parts makes the tests a bit too hard to grok but perhaps we could organize more around queries (instead of output data like transactions/segments) and then have sub tests that assert the expected pieces. Which maybe could mostly share implementation. A thought anyways. While we shouldn't rework transaction naming often, we def have a lot of similar details spread around even with existing shared test files. Made more complicated by the path/operation name used both for segments and transaction name.


carlo-808 (commented):
This all looks good. Looks like the toil was around adjusting the tests. 😄  
I think you need to update the README to remove comments about the deepest-path metric.

michaelgoin:
OK, documentation updates were quick so was able to squeeze those in before taking off. Hopefully caught everything.

docs/transactions.md
```diff
@@ -13,29 +13,96 @@ query {
...
+      issue,
+      title
+    }
+  }
+}
+```
+
+We will select a deepest unique path of 'libraries' as we select multiple fields beyond that point. Any resolver executed beyond that point may contribute to the performance characteristics of the transaction.
+
+If they query were to only select one field per resolver, we select the full path as each selection set is unique.
```
carlo-808, line 75:
```suggestion
If the query were to only select one field per resolver, we select the full path as each selection set is unique.
```

carlo-808 (approved):
Looks good. I see we will now use deepest unique path for transaction naming. Deepest path metrics have been removed and docs updated. 👍

Issue #138, opened 2021-10-15
Look into/fix test issues with latest gateway versions


michaelgoin:
Reference:

* https://github.com/newrelic/newrelic-node-apollo-server-plugin/pull/137

PR #140, opened 2021-10-20
Updates federation testing to always use latest 

michaelgoin:
## Proposed Release Notes

* Bumped `@newrelic/test-utilities` to ^6.1.1.

* Updated apollo-federation versioned tests to always test against latest dependency versions.

  Federated gateway modules are al < 1.0 releases and have incompatibilities with each other at various versions that impact our versioned testing. We now test only the latest pairs while these are in flight and may have various breaking changes. Once these go 1.0+, we can switch back to a permutation approach.

* Added `graphql` as a test dependency for apollo-federation versioned tests to fix issues on Node 12/14 runs.

* Limited samples to 15 for full versioned test runs to limit permutations as more versions are released.

## Links

* Fixes: https://github.com/newrelic/newrelic-node-apollo-server-plugin/issues/138

## Details

Looks like the TSD dep didn't update the notices previously for some reason so that got auto-jammed in here along with updating the test utils.

I'm guessing we'll eventually need `graphql` set for each of our test folders as it is a peer dependency but keeping this PR focused on resolving the federation test issues that currently exist.

### DIFF
--- a/THIRD_PARTY_NOTICES.md
+++ b/THIRD_PARTY_NOTICES.md
@@ -32,6 +32,7 @@ code, the source code can be found at [https://github.com/newrelic/newrelic-node
 * [prettier](#prettier)
 * [sinon](#sinon)
 * [tap](#tap)
+* [tsd](#tsd)
 
 
 ## dependencies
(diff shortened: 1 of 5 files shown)


michaelgoin:
Turns out I needed to cleanup node 16 modules in the versioned test folder to get to surface and is actually the result of an earlier theory I had. :)

bizob2828 (approved):
green builds are 👍🏻

PR #146, opened 2021-11-01
Fix federation test failures

michaelgoin:
## Proposed Release Notes

* Pinned `graphql` to a 15.x version for compatibility with Apollo libraries at ^15.

* Updated federation test setup to use `buildSubgraphSchema` per deprecation warnings.

## Details

Failure looks different locally but crux is that `graphql` now has a v16 and I had this setup to always grab latest.

Noticed a deprecation warning while I was at it and resolved.


bizob2828 (approved):

michaelgoin:
I've signed the CLA and work at NR. Merging (given approval) while the tool is having issues.

PR #148, opened 2021-11-11
fix: apollo plugins type mismatch,

newmind:
## Proposed Release Notes
For compatibility with 1.1.0, default function should accept non-argument


garbados (approved):
Thanks for your contribution! This looks good to me.

PR #153, opened 2021-11-30
Removes direct usage of internal tracer.

michaelgoin:
## Proposed Release Notes

* Removed direct usage of internal tracer.

## Links

* https://github.com/newrelic/node-newrelic/blob/main/lib/transaction/tracer/index.js#L48-L55
* https://github.com/newrelic/node-newrelic/blob/main/lib/shim/shim.js#L1232-L1238
* https://github.com/newrelic/node-newrelic/blob/main/lib/errors/error-collector.js#L190-L214


bizob2828 (approved):
:shipit:

PR #168, opened 2022-02-07
Fixes several dev-only dependency audit warnings.

michaelgoin:
## Proposed Release Notes

* Fixed several dev-only dependency audit warnings.


michaelgoin:
Looks like forcing the deps to update breaks some tests (or test assumptions).

Closing for now as this will take a bit more TLC and these are dev-only warnings.

PR #177, opened 2022-03-02
Updates tests to use latest version of newrelic 

michaelgoin:
## Proposed Release Notes

* Bumped `newrelic` dev dependency to ^8.8.0.

* Updated versioned testing to account for built-in agent instrumentation for Fastify.

## Details

This will enable running versioned tests from this repository in the main agent. Also makes the tests more representative of the current naming you'd see with up to date agents.

### DIFF
--- a/tests/versioned/apollo-server-fastify/segments.test.js
+++ b/tests/versioned/apollo-server-fastify/segments.test.js
@@ -13,6 +13,11 @@ const UNKNOWN_OPERATION = '<unknown>'
 const OPERATION_PREFIX = 'GraphQL/operation/ApolloServer'
 const RESOLVE_PREFIX = 'GraphQL/resolve/ApolloServer'
 
+// Varies based on apollo-server-fastify version.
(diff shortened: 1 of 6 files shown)


tests/versioned/apollo-server-fastify/segments.test.js
```diff
@@ -41,16 +46,22 @@ function createFastifySegmentsTests(t, frameworkName) {
           name: `${TRANSACTION_PREFIX}//${operationPart}`,
           children: [
             {
-              name: `${OPERATION_PREFIX}/${operationPart}`,
+              name: GQL_MIDDLEWARE_SEGMENT_PATTERN,
```
michaelgoin, line 49: Thought about adding a helper so wouldn't need to insert this on each item but decided against it.

Seemed like it would obscure understanding of the structure too much. Also, this captured structure isn't likely to change often.

bizob2828 (approved):

PR #267, opened 2023-07-26
feat: enhance typings for customResolverAttributes and customOperationAttributes

a88zach:
This change enhances the typings for customResolverAttributes and customOperationAttributes so that it enforces the value in the key/value pairs returned must be a string | number | boolean.  It also uses generics so that the consumer can correctly type the function parameters regardless if they are using v3 or v4 of Apollo Server


bizob2828 (approved):
@a88zach this is way more verbose. def flexing your TS muscles here 💪🏻

PR #353, opened 2025-07-17
chore: Pinned `apollo-server` test version to below 5.0.0

svetlanabrennan:
Temporarily pinning tests to below 5.0.0 for apollo-server until we fix tests.


svetlanabrennan:
Related to issue: https://github.com/newrelic/node-newrelic/issues/3235

